PR #3, opened 2017-12-16
Inline the onSuccess() and onFailure() functions

kevinherron:
It looks like all the other functions that receive callbacks are already inlined. These two were missed, and without being inlined the callbacks can't, for example, suspend the current coroutine.


michaelbull:
Thanks! I've also gone ahead and marked them as `infix` too (3e3b649193e58fb0707724fe260bcf38db54329a), which may improve usage in DSL environments.

PR #10, opened 2020-02-16
Add a toError method

Globegitter:
Allows to turn an Ok result into an Err given the provided predicate

This is something we now make use of quite often and is a bit nicer than always having to use a `flatMap`. Not sure if that first best into this file, but what do you think about it?

### DIFF
--- a/src/main/kotlin/com/github/michaelbull/result/Map.kt
+++ b/src/main/kotlin/com/github/michaelbull/result/Map.kt
@@ -166,3 +166,19 @@ inline infix fun <V, E, U> Result<V, E>.flatMap(transform: (V) -> Result<U, E>):
 
     return andThen(transform)
 }
+
+/**
+ * If this [Result] is [Ok] and the [predicate] matches then returns the new [error][Err.error] otherwise the 
+ * unchanged [Result<V, E>][Result]
+ *
+ */
+inline fun <V, E> Result<V, E>.toError(predicate: (V) -> Boolean, error: (V) -> E): Result<V, E> {
+    return when (this) {
+        is Ok -> if (predicate(value)) {
+            Err(error(value))
+        } else {
+            this
+        }
+        is Err -> this
+    }
+}


Globegitter:
If you think this makes sense or have another idea to achieve the same I will of course add some tests.

Globegitter:
@michaelbull friendly ping in case this notification got lost :smile:

Globegitter:
Also another one I have got now:

```
fun <V, E> Result<V?, E>.toError(error: E): Result<V, E> {
    return when (this) {
        is Ok -> value?.let { Ok(it) } ?: Err(error)
        is Err -> this
    }
}
```

sometimes it is not possible to use the exist `.toResultOr` method so it would be nice to have an easy way later on to convert a nullable result to a non-nullable one - what do you think?

michaelbull:
HI, thanks for the suggestions.

I'm definitely more interested in the one in the first comment of the PR, I think naming it something like `toErrorIf` would be more appropriate to indicate the existence of a predicate ([example of this in the Java stdlib](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#removeIf-java.util.function.Predicate-)).

I'll have a think about the implications of both and play with them on a local codebase to see how robust they are. Generally I'm aiming to keep the library lean but still provide enough utility to support the most common use cases, which these additions may well fall under.

I'll let you know when I've had a chance to put them to use - cheers.

Globegitter:
Yes that makes total sense and that is also what extension function are useful for, just let me know :)

michaelbull:
Hi,

Apologies for the delay.. I've rebased and modified your commit slightly into master.

I also added a `toErrorUnless` alongside the `toErrorIf`, this is because kotlin's stdlib has analogous methods for converting a value to null if a predicate is satisfied: [`takeIf`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/take-if.html), [`takeUnless`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/take-unless.html).

You can see the diff in the following commits: https://github.com/michaelbull/kotlin-result/compare/4d2a2d145311ea24ee164941f465178791e8f883...1000c588c01a6f1fa9133329c11681ecda112a95

Please let me know if you have any more comments, and if you'd be willing to write some unit tests that would be very much appreciated.

michaelbull:
Ping @Globegitter

PR #13, opened 2020-05-08
Add monad comprehensions via binding block with bind extension function

Munzey:
Hello!

Thought I would open a pull request to add monad comprehension like syntax for this result type.
Full disclosure: before i knew about this repo I had looked to adding this feature for result4k: https://github.com/npryce/result4k/pull/3
Wasn't getting a response there so I've moved on 😅 that's when I started looking around and found this very cool repo!

For that pr I did quite a bit of micro benchmarking with jmh. The main reason for not moving forward with arrow's either type for me is how poor performing it is. The apis are also very unstable (still not at a 1.0 release). So it was important for me to choose another lib that was performant.

With my (possibly contrived) test I have benchmarked and compared flatmapping versus using my proposed bind syntax for this repo. It came out a little slower: 

Lib | Success Flow | Failure Flow
------------ | ---------------- | -------------
ResultBinding | 187250 ops/ms | 178704 ops/ms
flatmap | 280932 ops/ms | 258156 ops/ms 

One thing to point out is that comparing this library to Result4k, the ops/ms is significantly slower in this repo for flatmapping (i get around 250,000 - 280,000 ops/ms in this repo, compared to almost 450,000 ops/ms in result4k on my machine). I suspect its related to the use of contracts and more object creation in this repo. Regardless, its still pretty fast! just thought its interesting.

Anyway, hope your open to the proposal and keeping well in these strange times!

### DIFF
--- a/src/main/kotlin/com/github/michaelbull/result/bind/BindFailure.kt
+++ b/src/main/kotlin/com/github/michaelbull/result/bind/BindFailure.kt
@@ -1,0 +1,9 @@
+package com.github.michaelbull.result.bind
+
+sealed class NoStackTraceException : Exception() {
+    override fun fillInStackTrace(): Throwable {
+        return this
+    }
+
+    object BindFailure : NoStackTraceException()
+}
--- a/src/main/kotlin/com/github/michaelbull/result/bind/BindFailure.kt
+++ b/src/main/kotlin/com/github/michaelbull/result/bind/BindFailure.kt
@@ -1,9 +1,8 @@
 package com.github.michaelbull.result.bind
 
-sealed class NoStackTraceException : Exception() {
+internal object BindFailure : Exception() {
+    // no stacktrace needed
     override fun fillInStackTrace(): Throwable {
         return this
     }
-
-    object BindFailure : NoStackTraceException()
 }
--- a/src/main/kotlin/com/github/michaelbull/result/bind/ResultBindingImpl.kt
+++ b/src/main/kotlin/com/github/michaelbull/result/bind/ResultBindingImpl.kt
@@ -4,7 +4,6 @@ import com.github.michaelbull.result.Err
 import com.github.michaelbull.result.Ok
 import com.github.michaelbull.result.Result
 import com.github.michaelbull.result.getOrElse
-import com.github.michaelbull.result.bind.NoStackTraceException.BindFailure
 
 /**
  * Calls the specified function [block] with [ResultBindingImpl] as its receiver and returns a [Result] of either [Ok] with value [V] or [Err] with error [E].
@@ -38,9 +37,8 @@ internal class ResultBindingImpl<E> : ResultBinding<E> {
     internal companion object {
         inline fun <V, E> with(crossinline block: ResultBindingImpl<E>.() -> V): Result<V, E> {
             val context = ResultBindingImpl<E>()
-            val wrapResult: ResultBindingImpl<E>.() -> Result<V, E> = { Ok(block()) }
             return try {
-                with(context, wrapResult)
+                with(context) { Ok(block()) }
             } catch (e: BindFailure) {
                 context.error
             }
--- a/src/main/kotlin/com/github/michaelbull/result/bind/ResultBindingImpl.kt
+++ b/src/main/kotlin/com/github/michaelbull/result/bind/ResultBindingImpl.kt
@@ -1,0 +1,49 @@
+package com.github.michaelbull.result.bind
+
+import com.github.michaelbull.result.Err
+import com.github.michaelbull.result.Ok
+import com.github.michaelbull.result.Result
+import com.github.michaelbull.result.getOrElse
+import com.github.michaelbull.result.bind.NoStackTraceException.BindFailure
+
+/**
+ * Calls the specified function [block] with [ResultBindingImpl] as its receiver and returns a [Result] of either [Ok] with value [V] or [Err] with error [E].
+ *
+ * [ResultBinding.bind] extension function is made available to any [Result] object inside this [block].
+ * [ResultBinding.bind] will attempt to unwrap the [Result] to return its [Ok] value locally.
+ * If the result is a [Err], then the [binding] block terminates with that [ResultBinding.bind] and returns a [Result] containing the error of the failed binding.
+ *
+ * This allows for easy unwrapping of result type objects along a happy path.
+ * for examples, see tests
+ * @sample com.github.michaelbull.result.bind.ResultBindingTest
+ *
+ */
+inline fun <V, E> binding(crossinline block: ResultBinding<E>.() -> V): Result<V, E> {
+    return ResultBindingImpl.with(block)
+}
+
+@PublishedApi
+internal class ResultBindingImpl<E> : ResultBinding<E> {
+
+    lateinit var error: Err<E>
+
+    override fun <V> Result<V, E>.bind(): V {
+        return this.getOrElse {
+            error = Err(it)
+            throw BindFailure
+        }
+    }
+
+    @PublishedApi
+    internal companion object {
+        inline fun <V, E> with(crossinline block: ResultBindingImpl<E>.() -> V): Result<V, E> {
+            val context = ResultBindingImpl<E>()
+            val wrapResult: ResultBindingImpl<E>.() -> Result<V, E> = { Ok(block()) }
+            return try {
+                with(context, wrapResult)
+            } catch (e: BindFailure) {
+                context.error
+            }
+        }
+    }
+}
--- a/src/main/kotlin/com/github/michaelbull/result/bind/ResultBinding.kt
+++ b/src/main/kotlin/com/github/michaelbull/result/bind/ResultBinding.kt
@@ -1,0 +1,7 @@
+package com.github.michaelbull.result.bind
+
+import com.github.michaelbull.result.Result
+
+interface ResultBinding<E> {
+    fun <V> Result<V, E>.bind(): V
+}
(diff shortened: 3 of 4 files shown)


michaelbull:
Wow this is a very interesting concept. Is this inspired from an existing Result/Either impl? If so I'd like to look over how it compares to an established implementation.

> One thing to point out is that comparing this library to Result4k, the ops/ms is significantly slower in this repo for flatmapping

This is particularly interesting. I wasn't aware that the compiler contracts would have any effect on performance?

Looking at [result4k's flatMap](https://github.com/npryce/result4k/blob/master/src/main/kotlin/com/natpryce/result.kt#L31), they are doing the same thing as in this library and not applying any fewer object allocations than I do. So I can't see any other explanation other than the inclusion of contracts.

#### result4k

```kotlin
inline fun <T, Tʹ, E> Result<T, E>.flatMap(f: (T) -> Result<Tʹ, E>): Result<Tʹ, E> =
    when (this) {
        is Success<T> -> f(value)
        is Failure<E> -> this
    }
```

#### kotlin-result

```kotlin
inline infix fun <V, E, U> Result<V, E>.andThen(transform: (V) -> Result<U, E>): Result<U, E> {
    contract {
        callsInPlace(transform, InvocationKind.AT_MOST_ONCE)
    }

    return when (this) {
        is Ok -> transform(value)
        is Err -> this
    }
}
```

src/main/kotlin/com/github/michaelbull/result/bind/BindFailure.kt
```diff
@@ -0,0 +1,9 @@
+package com.github.michaelbull.result.bind
+
+sealed class NoStackTraceException : Exception() {
+    override fun fillInStackTrace(): Throwable {
+        return this
+    }
+
+    object BindFailure : NoStackTraceException()
+}
```
michaelbull, line 9: Is there a reason we model the `NoStackTraceException` as a `sealed class` where the only impl is `BindFailure`?

I suspect this can be simplified down to:

```kotlin
object BindFailure : Exception() {
    override fun fillInStackTrace(): Throwable {
        return this
    }
}
```

src/main/kotlin/com/github/michaelbull/result/bind/ResultBindingImpl.kt
```diff
@@ -0,0 +1,49 @@
...
+ * [ResultBinding.bind] extension function is made available to any [Result] object inside this [block].
+ * [ResultBinding.bind] will attempt to unwrap the [Result] to return its [Ok] value locally.
+ * If the result is a [Err], then the [binding] block terminates with that [ResultBinding.bind] and returns a [Result] containing the error of the failed binding.
+ *
+ * This allows for easy unwrapping of result type objects along a happy path.
+ * for examples, see tests
+ * @sample com.github.michaelbull.result.bind.ResultBindingTest
+ *
+ */
+inline fun <V, E> binding(crossinline block: ResultBinding<E>.() -> V): Result<V, E> {
```
michaelbull, line 21: I think we should make this function name more specific, for example `resultBinding` or `bindResult`, as `binding` feels too generic. Let me know if you have a good proposal for this.

src/main/kotlin/com/github/michaelbull/result/bind/ResultBindingImpl.kt
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @PublishedApi
+    internal companion object {
+        inline fun <V, E> with(crossinline block: ResultBindingImpl<E>.() -> V): Result<V, E> {
+            val context = ResultBindingImpl<E>()
+            val wrapResult: ResultBindingImpl<E>.() -> Result<V, E> = { Ok(block()) }
+            return try {
+                with(context, wrapResult)
+            } catch (e: BindFailure) {
```
michaelbull, line 44: I think we can go ahead and inline the `wrapResult` argument here to avoid the nasty type declaration on L41:

```kotlin
return try {
    with(context) { Ok(block())) }
} catch(e: BindFailure) {
    ...
```

src/main/kotlin/com/github/michaelbull/result/bind/ResultBindingImpl.kt
michaelbull: Upon further inspection, it looks like a library called [`bow`](https://bow-swift.io/docs/patterns/monad-comprehensions/) for swift calls this function "binding", so let's keep it that way.

src/main/kotlin/com/github/michaelbull/result/bind/ResultBindingImpl.kt
Munzey: good call!

src/main/kotlin/com/github/michaelbull/result/bind/ResultBindingImpl.kt
Munzey: ah nice, yep bow is the counter part to arrow so this makes sense!

Munzey: but i agree, perhaps it is a bit too generic, and I agree `resultBinding` might be the best way to phrase what is is - my only issue is it becomes quite a long word to use for what is essentially a keyword.

src/main/kotlin/com/github/michaelbull/result/bind/BindFailure.kt
Munzey: no reason at all. I think at the time of creatign it I just wanted to be verbose about what everything was. But ye i think it makes sense to simplify it!

Munzey:
@michaelbull thanks for fast and detailed response! 🥰 

>Is this inspired from an existing Result/Either impl? If so I'd like to look over how it compares to an established implementation.

So this is the output of having investigated using arrow, then having gone to see this talk at kotlin conf last year: https://www.youtube.com/watch?v=pvYAQNT4o0I 
I came from writing scala before moving to kotlin and i really missed some of the language ability such as [for comprehensions](https://docs.scala-lang.org/tour/for-comprehensions.html)

Arrow added this functionality via [Monad Comprehensions](https://arrow-kt.io/docs/0.10/patterns/monad_comprehensions/).
Because they are emulating higher kinded types, the code is fairly complex in how they achieve this syntax. They also use coroutines to do this. If you were interested in their implementation I would take a look at the following files:
https://github.com/arrow-kt/arrow-core/blob/master/arrow-core/src/main/kotlin/arrow/core/extensions/either.kt#L256
https://github.com/arrow-kt/arrow-core/blob/master/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt#L31
https://github.com/arrow-kt/arrow-core/blob/master/arrow-core-data/src/main/kotlin/arrow/typeclasses/suspended/BindSyntax.kt

My intial attempt was to basically strip back their coroutine code to work strictly for result type (i mention the benchmark results of that in the description of https://github.com/npryce/result4k/pull/3).
Basically while it was a huge improvement over arrows implementation that covers all of their monadic types, I realised I didnt really need coroutines at all, so really all i ended up keeping was the clever scoped lambda type signature and the syntax convention. So without coroutines suddenly we see speeds comparable to just using the normal flatmap provided in these result libraries. 
So to answer the question in a long winded way 😅 no there isnt really a comparable implementation that I know of!

Munzey:
also regarding
>On a wider note, if you have done some benchmarking with JMH or similar then I am very open to adding a benchmark framework to this repository. Let me know if that's something you'd like to contribute to.

sure I can open a pull request with benchmarking added. SOmething I never tried to do actually was compare any of these libraries to the internal kotlin result type. Might be interesting!
Should mention that I'm no expert on micro benchmarking - so maybe when i put up the pr someone can chime in on whether there are better ways to measure.

src/main/kotlin/com/github/michaelbull/result/bind/ResultBindingImpl.kt
Munzey: interestingly if you go far back enough in the arrow repo (back to when it was called kategory) they had considered calling `bind` -> `await`. but might have been a good choice to not confuse this as being something async or with coroutines

michaelbull:
Added in 9bcaa974ca03b9f518a1e84717f79272f4d090c2

Munzey:
nice! really like the `when` pattern matching!

michaelbull:
I replaced the `this.getOrElse` with the `when` purely to avoid allocating an extra `Err` - instead we use the `Err` that it already is.

Munzey:
also I can rebase #14 now on to point to new master

Munzey:
@michaelbull do you think it would make sense to add this functionality to the readme? I can open a proposed change if you'd like? Up to you 😄

michaelbull:
Yeah, I think a section titled "monad comprehensions" or something would be good. If you could also add the links to the example implementations (e.g. swift's bow lib) to emphasise where the inspiration comes from that would be helpful.

Is there a way to get the benchmarking to output the results in a markdown formatted table? I would like to add benchmark results to the README (ran on my desktop) but don't see an easy way of displaying the data.

Munzey:
@michaelbull hmm the only output I know of is in `build/reports/benchmarks/main/[timestamp]/benchmark.json`

might need to write a script to generate some pretty output of that

michaelbull:
Okay, I found an online visualiser but it required me to put all the benchmarks in one class for them to be compared properly:

image

Munzey:
pretty cool!
find that graphic to be fairly biased though - while `bindingSuccess` is significantly less ops/ms, the x axis distances seem very strange 😅

michaelbull:
It's a logarithmic scale, you can still compare the raw numbers appropriately

PR #14, opened 2020-05-09
Benchmarking [Merge after pr #13]

Munzey:
This pr also contains the changes introduced in https://github.com/michaelbull/kotlin-result/pull/13

Adds benchmarking via https://github.com/Kotlin/kotlinx-benchmark
Under the hood this gradle plugin uses the [Java Microbenchmark Harness (JMH)](https://openjdk.java.net/projects/code-tools/jmh/)
A great guide which describes how to use jmh correctly can be found here: http://tutorials.jenkov.com/java-performance/jmh.html

I'm no expert at this. The benchmark tests themselves that I added are based on the work that @npryce had already done for result4k: https://github.com/npryce/result4k/blob/master/src/jmh/kotlin/com/natpryce/benchmarks.kt

I have added benchmarks for flatmap of this library and result4k, as well as included a benchmark for my proposed binding syntax. 
There would be nothing stopping someone adding tests for kittinunfs result type, or to compare kotlin's internal result type as well. Or even arrow's Either type if you're curious!

To run, simply call gradle task `benchmark`.
Additional configurations for running bench marks can be specified in build.gradle: https://github.com/Kotlin/kotlinx-benchmark#configuration


build.gradle.kts
```diff
@@ -30,13 +33,40 @@ allprojects {
...
     implementation(kotlin("stdlib"))
     testImplementation("junit:junit:4.13")
     testImplementation(kotlin("test-common"))
     testImplementation(kotlin("test-annotations-common"))
     testImplementation(kotlin("test-junit"))
     testImplementation(kotlin("test"))
+
+    benchmarkImplementation(sourceSets.main.get().output + sourceSets.main.get().runtimeClasspath)
+    benchmarkImplementation("org.jetbrains.kotlinx:kotlinx.benchmark.runtime-jvm:0.2.0-dev-7")
+    benchmarkImplementation("com.natpryce:result4k:2.0.0")
```
michaelbull, line 68: I don't really want to benchmark against other implementations at this stage, but rather provide some benchmarks as to how people should expect *this* library to perform, and allow us to benchmark our own solutions (e.g. binding vs flatMap).

settings.gradle.kts
```diff
@@ -1,3 +1,10 @@
 rootProject.name = "kotlin-result"
 
 include("example")
+
+pluginManagement {
+    repositories {
+        maven("https://dl.bintray.com/kotlin/kotlinx" )
+        gradlePluginPortal()
+    }
+}
```
michaelbull, line 10: Is this stuff necessary or can we get away with just the repository declarations in the `build.gradle.kts` files?

src/benchmark/kotlin/com/natpryce/FlatMapBenchmark.kt
```diff
@@ -0,0 +1,40 @@
+package com.natpryce
+
+import org.openjdk.jmh.annotations.*
```
michaelbull, line 3: Can we replace this wildcard import with more specific ones here please

build.gradle.kts
Munzey: that's fair enough!
I guess my reasoning was if you ever roadshow this library there will be the inevitable question from audience regarding how this fairs against the other implementations out there. Maybe those benchmarks belong in a separate repo however.
I'll remove that and the result4k benchmark from the pr

settings.gradle.kts
Munzey: seems to be the default way to configure repository location for gradle plugins now.
Gradle actually specifies that pluginManagement _must_ be specified inside settings.gradle

michaelbull:
Added in 0910e9ffe477a1daafbe1dd3fede452f2ae60bca

PR #17, opened 2020-05-17
andThenRun and andThenMap

f-lombardo:
Added `andThenRun` and `andThenMap` that are the equivalent of `andThen` and `map`  but using a lambda with receiver.

### DIFF
--- a/example/src/main/kotlin/com/github/michaelbull/result/example/service/CustomerService.kt
+++ b/example/src/main/kotlin/com/github/michaelbull/result/example/service/CustomerService.kt
@@ -1,7 +1,6 @@
 package com.github.michaelbull.result.example.service
 
-import com.github.michaelbull.result.Result
-import com.github.michaelbull.result.andThen
+import com.github.michaelbull.result.*
 import com.github.michaelbull.result.example.model.domain.Customer
 import com.github.michaelbull.result.example.model.domain.CustomerCreated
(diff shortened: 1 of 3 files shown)


michaelbull:
Hi, thanks for your contribution. These are the sort of extension methods that I would like to leave out of this library as there are infinitely many combinations of each function (having it as a single-arg lambda, having it as a receiver, etc).

I think I'd rather keep the library leaner and allow consumers to add their own extension functions with their own names. As you showed, they are really simple one liners but muddy the waters in terms of naming.

An example of the naming confusing occurs with `andThenMap`. Right now, the current `andThen` implementation is functionally equivalent to a function that exist in other libraries called `flatMap`. We actually provided our own function named `flatMap` which simply forwards to the more canonically named `andThen`. Therefore if we want to be consistent we would end up mirroring `andThenMap` with `flatMapMap` which doesn't make a great deal of sense when reading it. This is because `andThen` is already a mapping function, but your variant makes the transform argument a receiver and adds the name "map" on the end to distinguish it from the original `andThen`.

f-lombardo:
Hi,
thank you for your answer! That's really a matter of taste, and, as you are the creator of the library, your taste is the most valuable :-)
Thank you for creating this good piece of software.

michaelbull:
Thanks very much for your understanding @f-lombardo. If you do have other things you would like to contribute, you can always ask my opinion before spending your valuable time! Thanks again.

Issue #12, opened 2020-04-20
Kotlin multi-platform support


SrgGrch:
This question was already raised (here #5), but still, I am not able to use this library in the Multiplatform project. I think this is due to differences between the android library's build.gradle and the multiplatform one. In the MP library you need to specify `kotlin("multiplatform")` plugin and divide code into source sets (for that specific library you will need only common source set). You can read more about [https://kotlinlang.org/docs/tutorials/mpp/multiplatform-library.html](Multiplatform Kotlin library) at kotlintlang.org

I've tested this library with MP "structure" and it seems to work, so if you are interested in supporting MP I can create PR for this and you will check out code.

michaelbull:
Hi, thanks for raising this. Last time I tried I had trouble publishing to bintray with the MPP structure, however we've since moved to Maven Central this isn't a problem (as I've already achieved it in my [`kotlin-inline-logger`](https://github.com/michaelbull/kotlin-inline-logger) library).

I'll try and copy what I did in my other library and get it working with the MPP structure soon. Thanks.

joreilly:
Starting from Kotlin 1.4 it should be possible to effectively call suspend functions from iOS etc code....do you know if this could be used as return type for functions like this?

michaelbull:
Hi, I tried giving this a go but couldn't get it to play well with the benchmarking framework. Perhaps you can give it a go?

I also tend to wonder what is actually stopping you from using it in a mpp in its current state. It doesn't depend on the jvm specific version of the kotlin stdlib, so even though you are importing it as a jar from maven central it should still be compiled against the platform agnostic kotlin-stdlib.

Munzey:
I can try get this working ✋

Munzey:
@michaelbull i have the benchmarking framework now working with mpp. It seems to be very picky about how the source set can be registered. I will put up a pr later today with proposed mpp structure (other than getting benchmarking working its basically a copy paste of kotlin-inline-logger gradle script).

avently:
@michaelbull for now I'm unable to use the lib inside commonMain part of my application. Do you plan to support it too? Or maybe just to merge #18 if it works in this case?

michaelbull:
@Munzey has kindly added this functionality which is now released as part of [1.1.7](https://github.com/michaelbull/kotlin-result/releases/tag/1.1.7)

Skeptick:
[...]

PR #18, opened 2020-06-05
convert to mpp structure

Munzey:
Proposed solution for #12
Things that had to change:
- moved files to adhere to app structure by placing in commonMain and commonTest
- `Binding.kt` overriding stack trace can’t be used in kotlin native as exception doesn’t have this function. **_EDIT:_** now using expect/actual keywords with jvm implementation that still overrides stack trace
- ~`BindingTest.kt `was importing junit test instead of kotlin test~ **_EDIT:_** merged in separate pr #19
- example application now needed to specify that it was compiling against Jvm 1.8
- added a jvmBenchmark sourceset for benchmarks that required java plugin to be enabled for the mpp jvm target


build.gradle.kts
```diff
@@ -10,86 +9,115 @@ description = "A Result monad for modelling success or failure operations."
...
+}
+
 allprojects {
     repositories {
         mavenCentral()
         jcenter()
         maven("https://dl.bintray.com/kotlin/kotlinx")
     }
-
-    tasks.withType<KotlinCompile> {
```
Munzey, line 28: this is now handled via language settings in mpp kotlin extension below

build.gradle.kts
```diff
@@ -10,86 +9,115 @@ description = "A Result monad for modelling success or failure operations."
...
+        val jvmMain by getting {
+            dependencies {
+                implementation(kotlin("stdlib-jdk8"))
+            }
+        }
+
+        val jvmTest by getting {
+            dependencies {
+                implementation(kotlin("test-junit"))
+                implementation(kotlin("test"))
```
Munzey, line 91: should this be moved to commonTest?

build.gradle.kts
```diff
@@ -10,86 +9,115 @@ description = "A Result monad for modelling success or failure operations."
...
     repositories {
         mavenCentral()
         jcenter()
         maven("https://dl.bintray.com/kotlin/kotlinx")
     }
-
-    tasks.withType<KotlinCompile> {
-        kotlinOptions {
-            jvmTarget = "1.8"
-            freeCompilerArgs = listOf("-Xuse-experimental=kotlin.contracts.ExperimentalContracts")
```
Munzey, line 31: this is now handled via language settings in mpp kotlin extension below

src/commonMain/kotlin/com/github/michaelbull/result/Binding.kt
```diff
@@ -38,11 +38,7 @@ inline fun <V, E> binding(crossinline block: ResultBinding<E>.() -> V): Result<V
     }
 }
 
-internal object BindException : Exception() {
-    override fun fillInStackTrace(): Throwable {
-        return this
-    }
-}
```
michaelbull, line 45: Given this change, surely we need to change this to an `expect` and `actual` contract, i.e. `expect class BindException` to exist and on the JVM one explicitly override the `fillInStackTrace()` in an `actual class BindException`? Otherwise we lose the behaviour of stacktrace avoidance

build.gradle.kts
```diff
@@ -10,86 +9,115 @@ description = "A Result monad for modelling success or failure operations."
...
 allprojects {
     repositories {
         mavenCentral()
         jcenter()
         maven("https://dl.bintray.com/kotlin/kotlinx")
     }
-
-    tasks.withType<KotlinCompile> {
-        kotlinOptions {
-            jvmTarget = "1.8"
```
Munzey, line 30: this is now specified via `jvm().compilations.all` below

src/commonTest/kotlin/com/github/michaelbull/result/BindingTest.kt
```diff
@@ -1,6 +1,6 @@
 package com.github.michaelbull.result
 
-import org.junit.Test
```
michaelbull, line 3: Can we split this change up into a separate PR? This can be done in isolation to the general MPP migration and is a good change to have whether the mpp gets merged or not.

src/commonMain/kotlin/com/github/michaelbull/result/Binding.kt
Munzey: ah good thinking!

build.gradle.kts
```diff
@@ -10,86 +9,115 @@ description = "A Result monad for modelling success or failure operations."
...
-    testImplementation(kotlin("test-common"))
-    testImplementation(kotlin("test-annotations-common"))
-    testImplementation(kotlin("test-junit"))
-    testImplementation(kotlin("test"))
+kotlin {
+    jvm {
+        withJava()
+    }
+    sourceSets {
+        val commonMain by getting {
```
michaelbull, line 69: Can we make this use the `named` methods, just like I do here: https://github.com/michaelbull/kotlin-inline-logger/blob/master/build.gradle.kts#L46

Otherwise IntelliJ moans that these `val`s (commonMain, commonTest, etc) are unused variables.

Munzey: ill take a look, it made specifying that `jvmBenchmark dependsOn(jvmMain)` nice and readable

build.gradle.kts
michaelbull: I don't think so, see: https://github.com/michaelbull/kotlin-inline-logger/blob/master/build.gradle.kts#L73

src/commonTest/kotlin/com/github/michaelbull/result/BindingTest.kt
Munzey: sure thats a fair point!

build.gradle.kts
```diff
@@ -10,86 +9,115 @@ description = "A Result monad for modelling success or failure operations."
...
-dependencies {
-    implementation(kotlin("stdlib"))
-    testImplementation("junit:junit:4.13")
-    testImplementation(kotlin("test-common"))
-    testImplementation(kotlin("test-annotations-common"))
-    testImplementation(kotlin("test-junit"))
-    testImplementation(kotlin("test"))
+kotlin {
+    jvm {
+        withJava()
```
michaelbull, line 66: What does this withJava thing do? I don't remember needing it in my other mpp library

build.gradle.kts
```diff
@@ -10,86 +9,115 @@ description = "A Result monad for modelling success or failure operations."
...
+            artifact(javadocJar.get())
+        }
+    }
 
-val sourcesJar by tasks.registering(Jar::class) {
-    group = LifecycleBasePlugin.BUILD_GROUP
-    description = "Assembles a jar archive containing the main classes with sources."
-    archiveClassifier.set("sources")
-    from(sourceSets["main"].allSource)
+    sourceSets.all {
```
michaelbull, line 116: Surely this can be put inside the other `sourceSets` block?

```kotlin
sourceSets {
    all {
         ...
    }

    named("commonMain") {
        ...
    }

...
...
```

build.gradle.kts
michaelbull: Oh if you are actually using the `jvmMain` as a variable then ignore me!

build.gradle.kts
Munzey: this was the magic line that got benchmarking working, see my comment here: https://github.com/michaelbull/kotlin-result/pull/18/files#diff-392475fdf2bc320d17762ed97109a121R54-R55

michaelbull: Haha, I can see why I couldn't get benchmarking working then... nice find!

Munzey: without it, we don't get a target that the benchmark plugin will recognize

build.gradle.kts
Munzey: makes sense! 👍

src/commonMain/kotlin/com/github/michaelbull/result/Binding.kt
Munzey: done

build.gradle.kts
Munzey: done

build.gradle.kts
```diff
@@ -10,86 +9,115 @@ description = "A Result monad for modelling success or failure operations."
...
+        kotlinOptions {
+            jvmTarget = "1.8"
+        }
+    }
+
+    jvm {
+        mavenPublication {
+            artifact(javadocJar.get())
+        }
+    }
```
michaelbull, line 120: Can these two jvm blocks be merged?

Something like

```kotlin
    jvm {
        mavenPublication {
            artifact(javadocJar.get())
        }

        compilations.all {
            kotlinOptions.jvmTarget = "1.8"
        }
    }
```

Munzey: yes! ill move them into the same block as `withJava()`

Munzey: done

src/commonTest/kotlin/com/github/michaelbull/result/BindingTest.kt
Munzey: opened pr: https://github.com/michaelbull/kotlin-result/pull/19

michaelbull:
Hi @Munzey, don't want to leave you in silence but the PR looks good right now. I'm busy right now so will hopefully get some time this weekend to merge it and release. I tried last weekend to release the binding behaviour but I had connection issues with uploading artifacts to maven central. Apologies for trailing on this one.

Munzey:
@michaelbull No worries! 
Appreciate the update :relaxed:

michaelbull:
Hi @Munzey. Been thinking about this change a bit today. Given we introduced binding before MPP but now the binding is platform dependent (expects a platform-specific impl of that Exception object we spoke about), I guess that means that we would need to add an equivalent implementation for every platform we support - whereas before the binding stuff we could have technically supported every platform as none of the code was platform specific in any way. I may be misunderstanding this though?

I was also therefore thinking of splitting out the binding stuff into a separate subproject and therefore separate artifact, such that the main library is platform independent and the binding stuff is platform dependent. I'm not sure how I feel about this and it could get ugly pretty quickly, maybe I'm misunderstanding the situation though. Let me know your thoughts.

Munzey:
>whereas before the binding stuff we could have technically supported every platform as none of the code was platform specific in any way. I may be misunderstanding this though?

@michaelbull I think you are right yes.

>I was also therefore thinking of splitting out the binding stuff into a separate subproject and therefore separate artifact, such that the main library is platform independent and the binding stuff is platform dependent. I'm not sure how I feel about this and it could get ugly pretty quickly,

This does sound like it would become a pain to support both. Some questions around it too, would it be two dependencies to add to a project like `mpp.result-core` and `jvm.result-binding`? or is there a `mpp.result-core` and a `jvm.result-extended` that comprises everything? Looking at a project like arrow, I think it can become confusing when there is multiple artifacts required to pull all with names that seem to suggest they do something similar 😅 .

Its a shame we can't have an expected `BindException` with a default implementation in the common code. There is actually an open issue to add this to kotlin (which I will now be up voting haha): https://youtrack.jetbrains.com/issue/KT-20427?_ga=2.19951672.1617774961.1592496293-2122254457.1527159951

The alternative here is to just not override the stacktrace for the jvm, I decided to look for some articles to mention how much faster is it to not fill in the stack trace. I did find one: https://www.atlassian.com/blog/archives/if_you_use_exceptions_for_path_control_dont_fill_in_the_stac
that suggests it is quite beneficial when used as part of flow control as it is here. Not sure if the current benchmarking we have written could be used to give much of an indicator.

Going back to the original open issue, I actually wonder is their even added value right now for making the project mpp to support android. The only thing preventing me pulling this lib into an android project was to specify that I wanted kotlin to compile against 1.8. Thats all it took. So It doesn't really need to be an mpp project to support android as far as I understand (I would assume for a kotlin native project that builds for both iOS and android that you could specify deps that are only for the android side of code).

That then leaves mpp projects where kotlin code is to be cross compiled to run for that platform , which right now I can imagine would be JS or Swift. Which comes back to your point
>I guess that means that we would need to add an equivalent implementation for every platform we support 

I think its really up to you whether you think its worth it 😄 Personally I'm very much in favor of keeping binding as part of the main lib, and simply adding the cross platform support for languages as the issues are opened. To hedge our bets I would say if we added a swift `actual` and a js `actual` and added those to the gradle mpp builds, there probably wouldnt be many requests coming in.

TL;DR im not sure lol

avently:
Everyday I open this page to look at the status of merging process. That's because I really need functionality that this library gives but I can't use it without multiplatform support. So i'm just wondering, will this pr be merged in near future or it can be unmerged for months?

Munzey:
@avently can I ask what platforms you plan to support with your project?

avently:
@Munzey Android and Java right now, Linux/Windows/Mac/iOS later.

michaelbull:
Merged in bca344daafadbc7af42089b88c712ac1061298dd

avently:
@Munzey and @michaelbull thank you, guys!

PR #25, opened 2020-07-17
 Add support for suspendable functions to be called inside of binding block

### LINKED ISSUE #24: Binding doesn't support coroutines
I was excited to try out the `binding` block in our codebase but it doesn't support suspending functions. We currently use `Result` heavily with coroutines so this was a bit of a bummer. Is there any likelihood of this being added in the future? I'm not sure if it's out of scope for this project. I'm happy to raise a PR if you think it would be a worthwhile addition.

Munzey:
Fixes  #24

As there was slightly more involved than initially thought to get testing working, I thought it best to leave adding benchmarking to a separate pr.

I ended up choosing to implement it in a separate package, `com.github.michaelbull.result.coroutines` to keep the names the same. After some thought I figured it was the cleanest way given how long the keyword is. And if someone really wants to use both versions in the same class, they can prefix with the package full name.
That said, If its something you are against, lets look to changing it to a name that's easy/short to use.

Also spotted that ZipTest had been merged after the MPP structure change but not refactored to the new folder directory so i pulled that under the correct folder if you are wondering what that file renaming is doing here.


build.gradle.kts
```diff
@@ -91,6 +91,7 @@ kotlin {
             dependencies {
                 implementation(kotlin("test-common"))
                 implementation(kotlin("test-annotations-common"))
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-native:1.3.7")
```
Munzey, line 94: turns out that using kotlinx-coroutines-core-common does not provide a `runBlocking` method, needed for testing coroutines. There is no runBlocking for KotlinJS, but native provides an implementation for jvm and ios.

michaelbull, line 94: I was hoping we would be able to keep the coroutine dependencies test only, is there a reason we need to have an implementation dependency on it?

src/commonTest/kotlin/com/github/michaelbull/result/RunTest.kt
```diff
@@ -0,0 +1,8 @@
+package com.github.michaelbull.result
+
+import kotlinx.coroutines.CoroutineScope
+
+/**
+ * Workaround to use suspending functions in unit tests
```
Munzey, line 6: found this solution here: https://github.com/Kotlin/kotlinx.coroutines/issues/885#issuecomment-446586161

src/commonTest/kotlin/com/github/michaelbull/result/RunTest.kt
```diff
@@ -0,0 +1,8 @@
...
+ */
+expect fun runTest(block: suspend (scope : CoroutineScope) -> Unit)
```
michaelbull, line 8: I think you want [`runBlockingTest`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/run-blocking-test.html)

build.gradle.kts
michaelbull: I think we should use `runBlockingTest` from `org.jetbrains.kotlinx:kotlinx-coroutines-test` as a `testImplementation` dependency. 

[This](https://github.com/michaelbull/kotlin-coroutines-jdbc) project has an example of using the coroutines testing library

Munzey: that is only for jvm though

Munzey: this is test only, its added to `commonTest`

src/commonTest/kotlin/com/github/michaelbull/result/RunTest.kt
Munzey: yep see other comment, unfortunately thats only for jvm projects

src/commonTest/kotlin/com/github/michaelbull/result/RunTest.kt
michaelbull: Might be worth leaving that link in the comment in the code, so we know where it came from and if a fix comes out we can remove it

src/jvmTest/kotlin/com/github/michaelbull/result/RunTest.kt
```diff
@@ -0,0 +1,6 @@
+package com.github.michaelbull.result
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.runBlocking
+
+actual fun runTest(block: suspend (scope : CoroutineScope) -> Unit) = runBlocking { block(this) }
```
michaelbull, line 6: We should probably rename this function to `runBlockingTest`, purely for consistency with that jvm-specific library naming.

src/commonTest/kotlin/com/github/michaelbull/result/RunTest.kt
Munzey: done

src/jvmTest/kotlin/com/github/michaelbull/result/RunTest.kt
Munzey: good call, done.

michaelbull:
Another great contribution, thanks @Munzey 

bd7e1244b31e59e7d26479d7673d24716f70be0d

PR #27, opened 2020-08-04
add separate benchmark to compare the different implementations of binding

Munzey:
Basically shows that the binding function has same performance for both implementations (except of course if we use async in the coroutines version, which I added for reference).

### DIFF
--- a/src/jvmBenchmark/kotlin/com/github/michaelbull/result/SuspendBindingBenchmark.kt
+++ b/src/jvmBenchmark/kotlin/com/github/michaelbull/result/SuspendBindingBenchmark.kt
@@ -5,34 +5,28 @@ import org.openjdk.jmh.annotations.*
 import org.openjdk.jmh.infra.Blackhole
 import java.util.concurrent.TimeUnit
 
-@BenchmarkMode(Mode.Throughput)
-@OutputTimeUnit(TimeUnit.SECONDS)
-@Threads(Threads.MAX)
+@BenchmarkMode(Mode.AverageTime)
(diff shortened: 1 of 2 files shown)


Munzey:
I'll come back to this once #29 is resolved

src/jvmBenchmark/kotlin/com/github/michaelbull/result/SuspendBindingBenchmark.kt
```diff
@@ -0,0 +1,77 @@
...
+
+        private fun provideX(): Result<Int, Error> {
+            Thread.sleep(time)
+            return Ok(1)
+        }
+        private fun provideY(): Result<Int, Error> {
+            Thread.sleep(time)
+            return Ok(2)
+        }
+        private suspend fun suspendedProvideX(): Result<Int, Error> {
```
Munzey, line 67: I recall now that Roman recommends this should be named `provideX()` and the blocking version `provideXBlocking()`

michaelbull:
Hi @Munzey, are there any plans to return to this PR or can we close it for now?

Munzey:
hey @michaelbull yes was planning to come back to this, its just blocked by #38 which there is a pr up now #41.
Obv #41 hasn't gotten much attention lately so will see if I can get that moving - otherwise yeah maybe we look to close this out end of the week?

michaelbull:
Merged in e1fd6387e58fcdb24d09dbbde4bda47ada9c850c

Preview of benchmarks:

image

PR #28, opened 2020-08-04
Concurrent failing binds causes uncaught BindException

Munzey:
I was adding tests to check which error is returned when multiple binds fail for suspendable binding. I suspected that for async calls that finish closely there could be a race condition and you end up with the error from say, the second failed call instead of the first.

Turns out there is a much bigger problem with implementation for async suspendable functions:

```kt
runBlocking {
            val result = binding<Int, BindingError> {
                val x = async { success().bind() }
                val y = async { fails().bind() } // weird stuff happens here and an uncaught BindException occurs
                val z = async { fails().bind() } 
                x.await() + y.await() + z.await()
            }
}
```

Basically its not behaving with the different coroutines running concurrently. If you step through the test I've added `returnsFirstErrIfBindingFailed()` you'll see what I mean.
Struggled to google a solution to this that could contain this to inside `binding`.
[This thread](https://github.com/Kotlin/kotlinx.coroutines/issues/763#issuecomment-615202690) with an answer from Roman I thought would do the trick but still not working.
CompletableDeferred might also be a solution but not sure how to incorporate it for this case.

I tried a bunch of solutions. Mainly around seeing was there a hack I could implement creating a second `ResultBinding` implementation for `SuspendableBinding`. No luck. All out of ideas right now so hoping others can help jump on this 😄

### DIFF
--- a/src/commonMain/kotlin/com/github/michaelbull/result/coroutines/SuspendableBinding.kt
+++ b/src/commonMain/kotlin/com/github/michaelbull/result/coroutines/SuspendableBinding.kt
@@ -1,10 +1,10 @@
 package com.github.michaelbull.result.coroutines
 
-import com.github.michaelbull.result.BindException
 import com.github.michaelbull.result.Ok
+import com.github.michaelbull.result.Err
 import com.github.michaelbull.result.Result
 import com.github.michaelbull.result.ResultBinding
-import com.github.michaelbull.result.ResultBindingImpl
+import kotlinx.coroutines.CancellationException
 import kotlin.contracts.InvocationKind
 import kotlin.contracts.contract
 
@@ -16,11 +16,31 @@ suspend inline fun <V, E> binding(crossinline block: suspend ResultBinding<E>.()
         callsInPlace(block, InvocationKind.EXACTLY_ONCE)
     }
 
-    val receiver = ResultBindingImpl<E>()
+    val receiver = SuspendableResultBindingImpl<E>()
 
     return try {
         with(receiver) { Ok(block()) }
-    } catch (ex: BindException) {
-        receiver.error
+    } catch (ex: BindCancellationException) {
+        receiver.internalError
+    }
+}
+
+internal object BindCancellationException : CancellationException(null)
+
+@PublishedApi
+internal class SuspendableResultBindingImpl<E> : ResultBinding<E> {
+
+    lateinit var internalError: Err<E>
+
+    override fun <V> Result<V, E>.bind(): V {
+        return when (this) {
+            is Ok -> value
+            is Err -> {
+                if (::internalError.isInitialized.not()){
+                    internalError = this
+                }
+                throw BindCancellationException
+            }
+        }
     }
 }
--- a/src/commonTest/kotlin/com/github/michaelbull/result/coroutines/SuspendableBindingTest.kt
+++ b/src/commonTest/kotlin/com/github/michaelbull/result/coroutines/SuspendableBindingTest.kt
@@ -81,7 +81,7 @@ class SuspendableBindingTest {
 
         suspend fun provideZ(): Result<Int, BindingError> {
             delay(1)
-            return Ok(2)
+            return Err(BindingError)
         }
 
         runBlockingTest {
--- a/src/jvmTest/kotlin/com/github/michaelbull/result/coroutines/AsyncSuspendableBindingTest.kt
+++ b/src/jvmTest/kotlin/com/github/michaelbull/result/coroutines/AsyncSuspendableBindingTest.kt
@@ -1,0 +1,75 @@
+package com.github.michaelbull.result.coroutines
+
+import com.github.michaelbull.result.*
+import kotlinx.coroutines.delay
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlinx.coroutines.*
+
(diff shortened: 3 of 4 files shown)


src/commonTest/kotlin/com/github/michaelbull/result/coroutines/SuspendableBindingTest.kt
```diff
@@ -81,7 +81,7 @@ class SuspendableBindingTest {
 
         suspend fun provideZ(): Result<Int, BindingError> {
             delay(1)
-            return Ok(2)
+            return Err(BindingError)
```
Munzey, line 84: while this isnt an issue here in the non async version, thought it made more sense to do something similar here given the name of the test implies a possible second failed binding.

src/jvmTest/kotlin/com/github/michaelbull/result/coroutines/AsyncSuspendableBindingTest.kt
```diff
@@ -0,0 +1,75 @@
...
+
+class AsyncSuspendableBindingTest {
+
+    private sealed class BindingError {
+        object BindingErrorA: BindingError()
+        object BindingErrorB: BindingError()
+    }
+
+    @Test
+    fun returnsOkIfAllBindsSuccessful() {
```
Munzey, line 18: passes

src/jvmTest/kotlin/com/github/michaelbull/result/coroutines/AsyncSuspendableBindingTest.kt
```diff
@@ -0,0 +1,75 @@
...
+            assertTrue(result is Ok)
+            assertEquals(
+                expected = 3,
+                actual = result.value
+            )
+        }
+    }
+
+    @Test
+    fun returnsFirstErrIfBindingFailed() {
```
Munzey, line 44: failing 😢

michaelbull:
If we made `binding` wrap the operation in it's own CoroutineScope, as per [this](https://github.com/Kotlin/kotlinx.coroutines/issues/763#issuecomment-615270914) comment, would it help?

Munzey:
@michaelbull ok got something working - will post soon

Munzey:
@michaelbull so the trick was to force the job to cancel once an error is encountered. 
i was experimenting and tried
```
coroutineScope {
    this.cancel()
}
```
and i was finally seeing a differnet exception thrown (job cancellation exception)!
at this point i realised i could get rid of throwing bind exception entirely and just switch to directly throwing our own cancellation exception.

The one downside to this is that `CancellationException` requires pulling `kotlinx-coroutines-core-native` into `commonMain` 😬

src/commonMain/kotlin/com/github/michaelbull/result/coroutines/SuspendableBinding.kt
```diff
@@ -16,11 +16,31 @@ suspend inline fun <V, E> binding(crossinline block: suspend ResultBinding<E>.()
...
+@PublishedApi
+internal class SuspendableResultBindingImpl<E> : ResultBinding<E> {
+
+    lateinit var internalError: Err<E>
+
+    override fun <V> Result<V, E>.bind(): V {
+        return when (this) {
+            is Ok -> value
+            is Err -> {
+                if (::internalError.isInitialized.not()){
```
Munzey, line 39: ensures only the first failing bind to complete is reported as the error

michaelbull:
> The one downside to this is that `CancellationException` requires pulling `kotlinx-coroutines-core-native` into `commonMain`

We're going to need a different solution then, as I think it's unacceptable to make this library inherently dependent on kotlin-coroutines. Maybe we have to split out the coroutines behaviour to a separate subproject?

src/commonMain/kotlin/com/github/michaelbull/result/coroutines/SuspendableBinding.kt
Munzey: still not entirely sure if this the best way to guarantee this.
 given this change already requires coroutines core to be added, perhaps it would be better to use [`Mutex`](https://kotlinlang.org/docs/reference/coroutines/shared-mutable-state-and-concurrency.html#mutual-exclusion) for reading if its empty and writing to it.

Munzey:
> Maybe we have to split out the coroutines behaviour to a separate subproject?

yep sounds like its the only way then to ensure the correct behaviour for suspendable version of binding while not tying users of this lib to that dependency. 
ccing @ditn and @sottti as they may be interested in this

Munzey:
@michaelbull if we went that route, would you be willing to host the repo and publish it?
If so I'd happily open a pr to get this work moved over to it.

michaelbull:
By a separate sub-project I mean a project within the same Gradle project, i.e. the same repository

Munzey:
ah fair enough ! right that makes sense, just a separate folder with its own gradle build script?

michaelbull:
The `example` directory is already it's own subproject (but isn't published), so along the same lines as that subproject.

Here's the Gradle docs on the subject: https://docs.gradle.org/current/userguide/multi_project_builds.html

Munzey:
@michaelbull thanks for the link! I'm gonna look into putting up a separate pr to get this into a subproject then

Munzey:
closing this. to be solved via #29

PR #31, opened 2020-08-28
More Kotlin1.4 migration things

Munzey:
Added `explicitApi()` so publicly exposed api functions needed to be explicitly declared `public` and all required return types.

Removed redundant kotlinx and stdlib dependencies. To do that I actually had to bump coroutines to 1.3.9 (they didn't make that very clear from the changelist 🤔 ).

### DIFF
--- a/kotlin-result-coroutines/build.gradle.kts
+++ b/kotlin-result-coroutines/build.gradle.kts
@@ -15,8 +15,7 @@ kotlin {
 
         val commonMain by getting {
             dependencies {
-                implementation(kotlin("stdlib-common"))
-                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-native:${Versions.kotlinCoroutines}")
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:${Versions.kotlinCoroutines}")
                 implementation(project(":kotlin-result"))
             }
         }
@@ -25,7 +24,6 @@ kotlin {
             dependencies {
                 implementation(kotlin("test-common"))
                 implementation(kotlin("test-annotations-common"))
-                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-native:${Versions.kotlinCoroutines}")
             }
         }
 
@@ -33,7 +31,6 @@ kotlin {
             dependencies {
                 implementation(kotlin("test-junit"))
                 implementation(kotlin("test"))
-                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:${Versions.kotlinCoroutines}")
             }
         }
     }
--- a/kotlin-result/src/commonMain/kotlin/com/github/michaelbull/result/Result.kt
+++ b/kotlin-result/src/commonMain/kotlin/com/github/michaelbull/result/Result.kt
@@ -7,19 +7,19 @@ package com.github.michaelbull.result
  * - Haskell: [Data.Either](https://hackage.haskell.org/package/base-4.10.0.0/docs/Data-Either.html)
  * - Rust: [Result](https://doc.rust-lang.org/std/result/enum.Result.html)
  */
-sealed class Result<out V, out E> {
+public sealed class Result<out V, out E> {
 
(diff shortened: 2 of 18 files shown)


kotlin-result-coroutines/build.gradle.kts
```diff
@@ -15,8 +15,7 @@ kotlin {
 
         val commonMain by getting {
             dependencies {
-                implementation(kotlin("stdlib-common"))
-                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core-native:${Versions.kotlinCoroutines}")
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:${Versions.kotlinCoroutines}")
```
Munzey, line 18: seems you only have to declare it the once now in commonMain!

kotlin-result/src/commonMain/kotlin/com/github/michaelbull/result/Result.kt
```diff
@@ -32,10 +32,10 @@ sealed class Result<out V, out E> {
 /**
  * Represents a successful [Result], containing a [value].
  */
-class Ok<out V>(val value: V) : Result<V, Nothing>() {
+public class Ok<out V>(public val value: V) : Result<V, Nothing>() {
 
-    override fun component1() = value
-    override fun component2() = null
+    override fun component1(): V = value
+    override fun component2(): Nothing? = null
```
Munzey, line 38: `Nothing?` 😑

kotlin-result/src/commonMain/kotlin/com/github/michaelbull/result/Result.kt
```diff
@@ -48,17 +48,17 @@ class Ok<out V>(val value: V) : Result<V, Nothing>() {
...
 
 /**
  * Represents a failed [Result], containing an [error].
  */
-class Err<out E>(val error: E) : Result<Nothing, E>() {
+public class Err<out E>(public val error: E) : Result<Nothing, E>() {
 
-    override fun component1() = null
-    override fun component2() = error
+    override fun component1(): Nothing? = null
```
Munzey, line 60: had to do it again here... seems like such a smell

michaelbull:
Thanks @Munzey, but can we split this change into two commits?

We have two logically separate changes here:

1. Removal of unnecessary third party libraries which will reduce out footprint
2. Adoption of the explicit API mode

The former is more important and doesn't rely on the latter.

kotlin-result/src/commonMain/kotlin/com/github/michaelbull/result/Result.kt
michaelbull: Seems right to me. If you try and destructure an `Ok<Int>` as such:

`val (good, bad) = Ok<Int>(5)`

`good` will be typed to `Int` and `bad` to `null`. Since you can't actually use `Nothing` as a variable, the only suitable approach is for the destructuring to return `null` (it can't return a `Nothing`).

Munzey: ye I'm just hung up on the semantics of "nullable nothing"

Munzey:
dependencies change now in separate pr #32

kotlin-result/src/commonMain/kotlin/com/github/michaelbull/result/Result.kt
michaelbull: I suppose if we made it throw an exception then you could make it return `Nothing` instead of `Nothing?`... but I don't expect people will want destructuring a `Result` to throw anything

Munzey: yep I agree having it throw would be bad - I think its more of a fallout from having the compiler allow all types be nullable even when you have something like `Nothing` that semantically expresses not being able to hold something 😅 .
Im sure the kotlin lang guys thought about this at some point during its creation, guess the pros just outweighed the cons for something so practical like nullability 🤷‍♂️

PR #32, opened 2020-08-28
clean up dependencies for 1.4

Munzey:


kotlin-result/build.gradle.kts
```diff
@@ -52,9 +46,6 @@ kotlin {
         }
 
         val jvmMain by getting {
-            dependencies {
-                implementation(kotlin("stdlib-jdk8"))
-            }
         }
```
michaelbull, line 49: I think we can remove this whole lambda now, given its empty

Munzey: done

michaelbull:
620f4342203797ce934b62b1b31560ccf60f4aa7

PR #35, opened 2020-09-18
fix links

gregoryinouye:
some of the github links on the README appear to be broken. this PR fixes the broken links.


michaelbull:
Looks like leftovers from our migration to a multi-module project. Thanks!

PR #36, opened 2020-10-29
Add JS targets for Kotlin Multiplatform

DerYeger:
This PR adds two targets for Kotlin/JS, namely browser and nodejs.

Minor changes to the common tests were necessary to ensure that they can fail on the JS platform.
I also left out the ordering test for async suspendable binding due to the single threaded nature of JS.

### DIFF
--- a/kotlin-result-coroutines/src/commonTest/kotlin/com.github.michaelbull.result.coroutines/binding/SuspendableBindingTest.kt
+++ b/kotlin-result-coroutines/src/commonTest/kotlin/com.github.michaelbull.result.coroutines/binding/SuspendableBindingTest.kt
@@ -25,7 +25,7 @@ class SuspendableBindingTest {
             return Ok(2)
         }
 
-        runBlockingTest {
+        return runBlockingTest {
             val result = binding<Int, BindingError> {
                 val x = provideX().bind()
                 val y = provideY().bind()
@@ -52,7 +52,7 @@ class SuspendableBindingTest {
             return Ok(x + 2)
         }
 
-        runBlockingTest {
+        return runBlockingTest {
             val result = binding<Int, BindingError> {
                 val x = provideX().bind()
                 val y = provideY(x.toInt()).bind()
@@ -84,7 +84,7 @@ class SuspendableBindingTest {
             return Ok(2)
         }
 
-        runBlockingTest {
+        return runBlockingTest {
             val result = binding<Int, BindingError> {
                 val x = provideX().bind()
                 val y = provideY().bind()
@@ -117,7 +117,7 @@ class SuspendableBindingTest {
             return Ok(2)
         }
 
-        runBlockingTest {
+        return runBlockingTest {
             val result = binding<Int, BindingError> {
                 val x = provideX().bind()
                 val y = provideY().bind()
--- a/kotlin-result-coroutines/src/jsTest/kotlin/com.github.michaelbull.result.coroutines/binding/AsyncSuspendableBindingTests.kt
+++ b/kotlin-result-coroutines/src/jsTest/kotlin/com.github.michaelbull.result.coroutines/binding/AsyncSuspendableBindingTests.kt
@@ -44,7 +44,7 @@ class AsyncSuspendableBindingTest {
     }
 
     @Test
-    fun returnsAnyErrIfBindingFailed() {
+    fun returnsAnyErrIfBindingFailed(): dynamic {
         suspend fun provideX(): Result<Int, BindingError> {
             delay(1)
             return Ok(1)
@@ -60,7 +60,7 @@ class AsyncSuspendableBindingTest {
             return Err(BindingError.BindingErrorB)
         }
 
-        runBlockingTest {
+        return runBlockingTest {
             val result = binding<Int, BindingError> {
                 val x = async { provideX().bind() }
                 val y = async { provideY().bind() }
--- a/kotlin-result-coroutines/src/jsTest/kotlin/com.github.michaelbull.result.coroutines/binding/AsyncSuspendableBindingTests.kt
+++ b/kotlin-result-coroutines/src/jsTest/kotlin/com.github.michaelbull.result.coroutines/binding/AsyncSuspendableBindingTests.kt
@@ -8,6 +8,7 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.delay
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
 class AsyncSuspendableBindingTest {
@@ -44,14 +45,14 @@ class AsyncSuspendableBindingTest {
     }
 
     @Test
-    fun returnsAnyErrIfBindingFailed(): dynamic {
+    fun returnsFirstErrIfBindingFailed(): dynamic {
         suspend fun provideX(): Result<Int, BindingError> {
-            delay(1)
+            delay(10)
             return Ok(1)
         }
 
         suspend fun provideY(): Result<Int, BindingError.BindingErrorA> {
-            delay(2)
+            delay(20)
             return Err(BindingError.BindingErrorA)
         }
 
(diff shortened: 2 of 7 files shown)


kotlin-result-coroutines/src/commonTest/kotlin/com.github.michaelbull.result.coroutines/binding/SuspendableBindingTest.kt
```diff
@@ -25,7 +25,7 @@ class SuspendableBindingTest {
             return Ok(2)
         }
 
-        runBlockingTest {
+        return runBlockingTest {
```
michaelbull, line 28: Is this necessary if we have a `return` on line 52? (also applies to other added `return` statements)

kotlin-result-coroutines/src/jsTest/kotlin/com.github.michaelbull.result.coroutines/binding/AsyncSuspendableBindingTests.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.github.michaelbull.result.Ok
+import com.github.michaelbull.result.Result
+import com.github.michaelbull.result.coroutines.runBlockingTest
+import kotlinx.coroutines.async
+import kotlinx.coroutines.delay
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class AsyncSuspendableBindingTest {
```
Munzey, line 13: I think even if ordering doesn't matter due to JS, it would still be good to add the test here for a failing bind to document the actual behaviour.

kotlin-result/src/jsMain/kotlin/com/github/michaelbull/result/BindException.kt
```diff
@@ -0,0 +1,3 @@
+package com.github.michaelbull.result
+
+internal actual object BindException : Exception()
```
Munzey, line 3: sucks that we have to do this, if you haven't already can you upvote: https://youtrack.jetbrains.com/issue/KT-20427 ?

kotlin-result-coroutines/src/commonTest/kotlin/com.github.michaelbull.result.coroutines/binding/SuspendableBindingTest.kt
DerYeger: Unfortunately yes. Without it the tests would never fail with Kotlin/JS. Instead, an `UnhandledPromiseRejectionWarning` is being printed to the console (see https://youtrack.jetbrains.com/issue/KT-22228#focus=Comments-27-3964713.0-0).

michaelbull: That's disappointing

kotlin-result-coroutines/src/jsTest/kotlin/com.github.michaelbull.result.coroutines/binding/AsyncSuspendableBindingTests.kt
DerYeger: Absolutely. I added a test for a failing async bind.

kotlin-result/src/jsMain/kotlin/com/github/michaelbull/result/BindException.kt
DerYeger: Of course, that's a great suggestion!

kotlin-result-coroutines/src/jsTest/kotlin/com.github.michaelbull.result.coroutines/binding/AsyncSuspendableBindingTests.kt
```diff
@@ -0,0 +1,74 @@
...
+
+        runBlockingTest {
+            val result = binding<Int, BindingError> {
+                val x = async { provideX().bind() }
+                val y = async { provideY().bind() }
+                val z = async { provideZ().bind() }
+                x.await() + y.await() + z.await()
+            }
+
+            assertTrue(result is Err)
```
Munzey, line 71: thanks for adding 😎  
Now I'm curious, is there just no guarantee which we will receive first in this case on JS? or will it always be `BindingErrorA` due to the order the `await`s are invoked? If so, I think you could be more precise in documenting the behaviour here.
e.g
"given multiple async failing binds, if invoked asyncronously, then the first invoked will always be the first error returned"

DerYeger: It seems that there's no guarantee for order. I ran the original test multiple times and had both successful and failed runs.

Munzey: thanks for checking. That's very interesting! So I guess that means on JS some order to execute the promises is set at runtime, and regardless of the delays, the single threaded execution just means thats the order they are completed in. 🤷‍♂️

kotlin-result-coroutines/src/jsTest/kotlin/com.github.michaelbull.result.coroutines/binding/AsyncSuspendableBindingTests.kt
```diff
@@ -0,0 +1,74 @@
...
+            delay(2)
+            return Err(BindingError.BindingErrorA)
+        }
+
+        suspend fun provideZ(): Result<Int, BindingError.BindingErrorB> {
+            delay(1)
+            return Err(BindingError.BindingErrorB)
+        }
+
+        runBlockingTest {
```
Munzey, line 63: does this not need a `return` as well? would that have affected the results discusssed below?

DerYeger: Ah yes, my bad. Should be fixed now.

kotlin-result-coroutines/src/jsTest/kotlin/com.github.michaelbull.result.coroutines/binding/AsyncSuspendableBindingTests.kt
Munzey: this made me realise, @michaelbull we should probably be verifying even for the simpler `BindingTest.kt` which binded functions will actually execute in the fail-to-bind test cases - i.e. it would be good to know if side effecting code in a binded function actually ran or not, and that the tests would verify the behaviour.

So in the case of `BindingTest.kt`, we know that no code after the failing bind will execute. So we should verify that. Likewise in `SuspendableBindingTest.kt`.
However in the case of `AsyncSuspendableBindingTest.kt` on jvm, I'm trying to recall whether all the async calls would still run to completion, or whether the cancelaltionException means the running coroutine is just dropped outright. Either way would be nice to have this confirmed by the test. 
Same applies to using async binding JS; we should verify if we expect all the binding code to finish or not, or whether its nondeterministic.

Munzey: referenced in https://github.com/michaelbull/kotlin-result/pull/37

Munzey: @DerYeger given the latest updates in #37 , could you add a test to this class to verify that on a bind failing, the other functions are cancelled/stop executing? See the state change test i added on the other pr for reference

DerYeger: @Munzey Sure! That will cause failing checks until the other PR is merged.

On another note, I retested the execution order with higher delays (10ms, 20ms and 1ms) and got different results. It seems that the the small delays of the original test (1ms, 2ms and 1ms) caused issues with scheduling in Kotlin/JS.

Munzey: very nice!
Ha nice that you spotted that. I actually found the same thing could happen on the first recompile of my tests on the other pr when using just a single ms in the difference - I also bumped the ms delay because of that. Not sure if there is an ideal way to test these types of discrepancies. Really just a byproduct of working with asynchronous code I guess.

Munzey (approved):
LGTM

PR #37, opened 2020-10-30
Eagerly cancel async bindings

Munzey:
After some of the discussion here: https://github.com/michaelbull/kotlin-result/pull/36#discussion_r514505228
I looked into verifying whether all binds will finish executing when ran asynchrounously if a failed binding is encountered and how we should verify the behaviour in our tests. I realised it could be useful to have for compute heavy binding functions ran asynchronously to be able to cancel them eagerly.

Just a suggestion, let me know what you think.

### DIFF
--- a/kotlin-result-coroutines/src/commonMain/kotlin/com/github/michaelbull/result/coroutines/binding/SuspendableBinding.kt
+++ b/kotlin-result-coroutines/src/commonMain/kotlin/com/github/michaelbull/result/coroutines/binding/SuspendableBinding.kt
@@ -3,7 +3,10 @@ package com.github.michaelbull.result.coroutines.binding
 import com.github.michaelbull.result.Err
 import com.github.michaelbull.result.Ok
 import com.github.michaelbull.result.Result
-import kotlinx.coroutines.*
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlin.contracts.InvocationKind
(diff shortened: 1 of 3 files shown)


kotlin-result-coroutines/src/commonMain/kotlin/com/github/michaelbull/result/coroutines/binding/SuspendableBinding.kt
```diff
@@ -12,18 +12,23 @@ import kotlin.contracts.contract
 /**
  * Suspending variant of [binding][com.github.michaelbull.result.binding].
  */
-public suspend inline fun <V, E> binding(crossinline block: suspend SuspendableResultBinding<E>.() -> V): Result<V, E> {
+public suspend inline fun <V, E> binding(eagerlyCancel: Boolean = false, crossinline block: suspend SuspendableResultBinding<E>.() -> V): Result<V, E> {
```
michaelbull, line 15: I'm not a fan of the boolean parameter. I would rather put the new behaviour in a separate method, e.g. `eagerBinding` or something.

kotlin-result-coroutines/src/commonMain/kotlin/com/github/michaelbull/result/coroutines/binding/SuspendableBinding.kt
```diff
@@ -3,7 +3,7 @@ package com.github.michaelbull.result.coroutines.binding
 import com.github.michaelbull.result.Err
 import com.github.michaelbull.result.Ok
 import com.github.michaelbull.result.Result
-import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.*
```
michaelbull, line 6: Need to revert this wildcard import

kotlin-result-coroutines/src/commonTest/kotlin/com.github.michaelbull.result.coroutines/binding/SuspendableBindingTest.kt
```diff
@@ -100,6 +101,90 @@ class SuspendableBindingTest {
...
+            return Err(BindingError)
+        }
+
+        suspend fun provideZ(): Result<Int, BindingError> {
+            delay(1)
+            zStateChange = true
+            return Err(BindingError)
+        }
+
+        runBlockingTest {
```
michaelbull, line 169: I guess we need a `return` statement on this line given our discussion in the most recent PR.

kotlin-result-coroutines/src/jvmTest/kotlin/com/github/michaelbull/result/coroutines/binding/AsyncSuspendableBindingTest.kt
```diff
@@ -76,4 +77,88 @@ class AsyncSuspendableBindingTest {
...
+            return Err(BindingError.BindingErrorA)
+        }
+
+        suspend fun provideZ(): Result<Int, BindingError.BindingErrorB> {
+            delay(1)
+            zStateChange = true
+            return Err(BindingError.BindingErrorB)
+        }
+
+        runBlocking {
```
michaelbull, line 104: Also here re: `return` statement.

kotlin-result-coroutines/src/jvmTest/kotlin/com/github/michaelbull/result/coroutines/binding/AsyncSuspendableBindingTest.kt
```diff
@@ -76,4 +77,88 @@ class AsyncSuspendableBindingTest {
...
```
michaelbull, line 146: And here

kotlin-result-coroutines/src/commonMain/kotlin/com/github/michaelbull/result/coroutines/binding/SuspendableBinding.kt
Munzey: ye I think that makes sense - it will probably help with testing too

michaelbull:
Just trying to wrap my head around this. So the existing behaviour has no gaurantee that if one fails the others will cancel? Or is my current understanding of the existing behaviour wrong?

kotlin-result-coroutines/src/commonTest/kotlin/com.github.michaelbull.result.coroutines/binding/SuspendableBindingTest.kt
Munzey: sure, If the other pr is merged first I will pull and fix these, otherwise maybe better to do in that pr.

Munzey:
>Just trying to wrap my head around this. So the existing behaviour has no gaurantee that if one fails the others will cancel? Or is my current understanding of the existing behaviour wrong?

I think the two new tests in `AsyncSuspendableBindingTest.kt` help clarify the most.

If your binding function contains 
```kt
{
  async{functionA.bind()}
  async{functionB.bind()}
}
```  
where functionB is some long running suspendable function and `functionA` fails to bind, then with:

`binding` - `functionB` is guaranteed to continue to run till the async job completes (this is the existing logic).
`eagerlyCancelBinding` - `functionB` will stop executing at next suspend point.

Issue #39, opened 2020-11-27
Mapping a failure to success given a predicate


oddsund:
While wrapping calls to an external api, one pattern emerged - recoverIf. So it would be similar to `toErrorIf`  and `toErrorUnless`, and would allow us to selectivly recover from known errors.

Is there any reason that this is not implemented? If not, I'll try to provide a PR during the weekend.

michaelbull:
Sounds like a good idea, would be willing to accept a PR for this 👍

PR #40, opened 2020-11-28
Add recoverIf and recoverUnless

oddsund:
These are similar to [toErrorIf] and [toErrorUnless], but maps from Err to Ok.

Solves #39.


kotlin-result/src/commonMain/kotlin/com/github/michaelbull/result/Or.kt
```diff
@@ -55,3 +55,43 @@ public inline infix fun <V, E> Result<V, E>.recover(transform: (E) -> V): Ok<V>
...
+            Ok(transform(error))
+        } else {
+            this
+        }
+    }
+}
+
+/**
+ * Returns the [transformation][transform] of the [error][Err.error] if this [Result] is [Err]
+ * and _does not_ satisfies the given [predicate], otherwise this [Result].
```
michaelbull, line 81: I think this should be "does not **satisfy**".

Issue #38, opened 2020-11-24
Move benchmarking to separate sub project


Munzey:
I tried to add a benchmark test for suspendable binding but ended up with cyclical dependencies trying to have `kotlin-result` `jvmBenchmark` source set depend on `kotlin-result-coroutine` project (which already depends on `kotlin-result`).

Would make life a lot easier to just move benchmarking to its own subproject. It would also simplify the current kotlin-result gradle file which has a bunch of tricks in it already just to get the benchmarking plugin to play nice inside an mpp build.

This would also help delineate dependencies if we decided to benchmark against other implementations (arrow's Either, kotlin's own result type,  kittinunf's result type).

Once completed we can look to move the work in #27 to there.

PR #41, opened 2020-11-28
Move benchmarks to separate module

hanleyt:
PR for #38.

I'm new to kmp libraries and was doing some guessing with the source set config. Let me know if this is what you were thinking @Munzey!


michaelbull:
This looks great. Massively simplifies the main subprojects Gradle configuration too. Ping @Munzey for a review but LGTM 👍

kotlin-result/build.gradle.kts
```diff
@@ -3,21 +3,6 @@ description = "A Result monad for modelling success or failure operations."
...
-    id("org.jetbrains.kotlin.plugin.allopen")
-    id("kotlinx.benchmark")
-}
-
-allOpen {
-    annotation("org.openjdk.jmh.annotations.State")
-    annotation("org.openjdk.jmh.annotations.BenchmarkMode")
-}
-
-sourceSets.create("benchmark")
```
Munzey, line 15: see this thread from when I was converting this build to mpp originally: https://github.com/michaelbull/kotlin-result/pull/18#discussion_r436006070

I recall running into issues trying to get the benchmark plugin to register any benchmark source when it existed inside the mpp project. 
Defining the "jvmBenchmark" source set alone wasn't enough. I believe I tried to create it as a custom source set inside of the mpp plugin: https://kotlinlang.org/docs/reference/mpp-dsl-reference.html#custom-source-sets. But that nothing was being registered as a target by the benchmark plugin. The multiplatform plugin turned out to provide a sort of workaround which may not be needed anymore. By calling `withJava()` inside the mpp `jvm` block, java code in the jvmBenchmark sourceset was magically picked up when registered to the benchmark plugin.

So you can definitely remove `withJava()` from this file, as well as `val jvmMain by getting` as that was only needed for the jvmBenchmark dependencies.

but perhaps some experimentation is needed to see what's the minimum needed to get this working for the new file

benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+    id("kotlinx.benchmark") version Versions.kotlinBenchmark
+    id("org.jetbrains.kotlin.plugin.allopen") version Versions.kotlin
+}
+
+allOpen {
+    annotation("org.openjdk.jmh.annotations.State")
+    annotation("org.openjdk.jmh.annotations.BenchmarkMode")
+}
+
+sourceSets.create("benchmark")
```
Munzey, line 14: as mentioned, this can go

benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+
+benchmark {
+    targets {
+        register("jvmBenchmark")
+    }
+}
+
+kotlin {
+    jvm {
+        withJava()
```
Munzey, line 24: ditto, can remove this.

benchmarks/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+allOpen {
+    annotation("org.openjdk.jmh.annotations.State")
+    annotation("org.openjdk.jmh.annotations.BenchmarkMode")
+}
+
+sourceSets.create("benchmark")
+
+benchmark {
+    targets {
+        register("jvmBenchmark")
```
Munzey, line 18: so at the time of converting to mpp, we only targeted jvm. But now there is a js target, and as we can see from the benchmark plugin description: https://github.com/Kotlin/kotlinx-benchmark
this should work for javascript too.

could we reorganize and move the benchmark code into `commonBenchmark`? 
I dont believe there is really anything in the current benchmark file that is jvm specific.

They have a great example for a multiplatform benchmark in the plugin repo, 
we can see the imports we would need to change here to make them multiplatform compatible: https://github.com/Kotlin/kotlinx-benchmark/blob/master/examples/kotlin-multiplatform/src/commonMain/kotlin/CommonBenchmark.kt

the build script for that example is also helpful: https://github.com/Kotlin/kotlinx-benchmark/blob/master/examples/kotlin-multiplatform/build.gradle

Munzey (changes requested):
Thanks again for taking a look at this!
Couple of suggestions

michaelbull:
Hi @Munzey & @hanleyt, can you shed some light on what the status of this PR is? Is it ok to merge or do we need to make some fixes before?

Munzey:
@michaelbull it would be great to just get some of the minor comments cleaned up before merging.
Although the switch to a `commonBenchmark` could be moved to a separate issue.
@hanleyt if your busy I got some time this week I can finish this pr if you like?

Munzey:
This can now be closed as it is continued in #47

PR #47, opened 2021-02-08
Move benchmarks to separate module

Munzey:
Follow on from  #41

Made the benchmark class fully multiplatform compliant and added the js target.
Unfortuantely the benchmarking plugin hasda couple of weird issues to get the js target working


benchmarks/src/commonMain/kotlin/com/github/michaelbull/result/BindingBenchmark.kt
```diff
@@ -1,16 +1,17 @@
...
+import kotlinx.benchmark.BenchmarkMode
+import kotlinx.benchmark.Mode
+import kotlinx.benchmark.OutputTimeUnit
+import kotlinx.benchmark.BenchmarkTimeUnit
+import kotlinx.benchmark.Benchmark
+import kotlinx.benchmark.Blackhole
+import kotlinx.benchmark.State
+import kotlinx.benchmark.Scope
 
+@State(Scope.Benchmark)
```
Munzey, line 12: For some reason, for js benchmark target to pick up any tests it needs the whole class annotated, opened a ticket: https://github.com/Kotlin/kotlinx-benchmark/issues/29

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -1,6 +1,6 @@
 object Versions {
     const val dokka = "0.10.1"
-    const val kotlin = "1.4.21-2"
+    const val kotlin = "1.4.10"
```
Munzey, line 3: js benchmarking failing to compile with anything higher, opened a ticket: https://github.com/Kotlin/kotlinx-benchmark/issues/28

michaelbull:
Can we try disabling the JS benchmarks instead of downgrading Kotlin to satisfy them being broken?

buildSrc/src/main/kotlin/Versions.kt
Munzey: disabled js benchmark target and bumped to 1.4.30

michaelbull:
Merged in 0df4c62d4f755d8e60197ad46d28bbf658c8173f

PR #48, opened 2021-03-24
Added Linux, Windows and MacOS targets

avently:
Made more native targets. Renamed iosMain into nativeMain which allows to have one directory for all native targets.
`./gradlew check` finishes successfully.

I don't have Macos so can't say it is working or not. If you have one, please, check that ios and macos targets work ok.
Same with the GitHub action. It should work but i don't have an account in publishing services to check.

Let me know if you find a bug in the implementation.


michaelbull:
Hi @avently, thanks for the PR!

Looking at the [GitHub checks for this PR](https://github.com/michaelbull/kotlin-result/pull/48/checks?check_run_id=2188973245), the `check` phase actually isn't being ran for Windows. This is because it's simply doing `./gradlew check`, which works under Bash on macOS and Ubuntu, but not on Windows. We instead need to do `gradlew.bat check` specifically on Windows.

Also, currently the macOS environment only runs test for iOS, but with the inclusion of the macOS native artifacts we should be able to add `maosX64Test` to [this line](https://github.com/michaelbull/kotlin-result/blob/master/.github/workflows/ci.yaml#L37) in the workflow.

> A problem I see in the implementation is that jvm and js artifacts will be built and uploaded ones per run (three times).

This is correct. In the [previous issue thread](https://github.com/michaelbull/kotlin-result/pull/23#issuecomment-771093966), there was an [example linked](https://github.com/LouisCAD/Splitties/blob/main/.github/workflows/release.yml) that published only the Mingx64 artifacts on Windows, the LinuxX64 artifacts on Ubuntu, and everything else on macOS. I think this is a good approach and the one I would recommend to solve this problem.

avently:
>the check phase actually isn't being ran for Windows. This is because it's simply doing ./gradlew check

That's because `./gradlew check` is allowed to run on ubuntu-latest only. That's intentional. Do we actually need to make platform-dependant checks? The lib doesn't have platform-dependant code (is has one nativeMain for every native platform). Do you think I should enable check for every platform?

>there was an example linked that published only the Mingx64 artifacts on Windows, the LinuxX64 artifacts on Ubuntu, and everything else on macOS

Thanks, I'll look into it.

michaelbull:
>  Do we actually need to make platform-dependant checks? The lib doesn't have platform-dependant code (is has one nativeMain for every native platform). Do you think I should enable check for every platform?

I think as a general rule of thumb, we should run as many tests on as many platforms that we publish & claim to support. I would feel uneasy adopting a library that has platform-specific artifacts that haven't passed unit tests on that platform.

avently:
Added checks for every target since it costs nothing. However it takes some time which can lead to spending of your free quota.
Force pushed with a Github email specified
Added upload of specific artefacts from specific targets. Js, Jvm, iosX64, iosArm64, macosX64 from Mac, mingwX64 from Windows and linuxX64 from Linux.

Now the result looks good! As you can see, for Windows I need to wrote `./gradlew` instead of `gradlew.bat`.

Is everything fine?

michaelbull:
Thanks @avently - I'll have a go at publishing the snapshots on the weekend and see if all artifacts come out as expected

michaelbull:
Looks like [everything is publishing](https://oss.sonatype.org/content/repositories/snapshots/com/michael-bull/kotlin-result/) - can you give it a look over @avently and confirm that all these artifacts are as expected?

avently:
@michaelbull their names are expected. Linux target can be imported and used in IDEA (I added sonatype snapshots into gradle sources and removed local build of the lib). Can't check ios and Mac's builds since don't have a mac. 
Mingw and macos artifacts were downloaded by gradle too when I specified macosX64 and mingwx64 targets.
So I think everything is fine. iUsers will tell us if something goes wrong with their gradle projects.

avently:
Can I close it since you already cherry-picked the commit?
And, by the way, what's the point of cherry-picking instead of merging a PR?

michaelbull:
Yep thanks for the confirmation. Looks like we will be fully multi-platform next release then, but hopefully the snapshot is enough for you to play with until I make a release this weekend. Also it wasn't cherry-picked, it was rebased using git's CLI instead of merging via GitHub.

Issue #50, opened 2021-03-25
Move coroutine tests to use kotlinx-coroutines-test instead of runBlockingTest


Munzey:
See Roman's latest comment: https://youtrack.jetbrains.com/issue/KT-22228#focus=Comments-27-4791713.0-0

michaelbull:
Also may be relevant: https://blog.jetbrains.com/kotlin/2021/04/kotlin-1-5-0-rc-released/#test-library-changes

michaelbull:
For some reason, the `kotlinx-coroutines-test` library depends on the JVM implementation of kotlin-coroutines, so it isn't cross-platform and thus can't replace the `runBlockingTest` in this library.

michaelbull:
Going to close this issue until https://github.com/Kotlin/kotlinx.coroutines/issues/1996 is resolved.

PR #57, opened 2021-06-15
resolve flaky jvm multi threaded coroutine tests

Munzey:
By switching to using the jvm only (hopefully by 1.6 in common) [coroutine test lib](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/) we can swap to using a fake dispatcher instead of defining threads which should resolve any odd timing issues that we've seen occur in some ci builds (e.g https://github.com/michaelbull/kotlin-result/runs/2810204490?check_suite_focus=true ).

Couple things to talk about in the pr regarding adopting the jvm only coroutine test lib:
- we discussed it a bit in #50, but it is a bit of a pain that we now have inconsistency between the `launch/async` tests added for for js and jvm platforms (though technically they were already out of sync when the additional tests for `launch` were added in #46). afaik, I think we are still prone to the timing issues for the js tests that exist, maybe we leave them as is unless we encounter a flaky case, cc @DerYeger on this one 🤔 
- we now use two functions named `runBlockingTest`, maybe we want to rename one (could import the test lib one as something different
- on using a different Dispatcher per coroutine - I believe I originally added having each start on a separate thread to guarantee that each was definitely using a separate thread (perhaps someone can clarify but I believe using jvms `runBlocking` all happens on main thread which was my justification). Now that we are using dummy TestDispatchers, I went with just copying the old behaviour, but perhaps we could just use a single one per test 🤷‍♂️

### DIFF
--- a/kotlin-result-coroutines/build.gradle.kts
+++ b/kotlin-result-coroutines/build.gradle.kts
@@ -31,6 +31,7 @@ kotlin {
             dependencies {
                 implementation(kotlin("test-junit"))
                 implementation(kotlin("test"))
+                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:${Versions.kotlinCoroutinesTest}")
             }
         }
 
--- a/kotlin-result-coroutines/src/jvmTest/kotlin/com/github/michaelbull/result/coroutines/binding/AsyncSuspendableBindingTest.kt
+++ b/kotlin-result-coroutines/src/jvmTest/kotlin/com/github/michaelbull/result/coroutines/binding/AsyncSuspendableBindingTest.kt
@@ -3,19 +3,19 @@ package com.github.michaelbull.result.coroutines.binding
 import com.github.michaelbull.result.Err
 import com.github.michaelbull.result.Ok
 import com.github.michaelbull.result.Result
-import kotlinx.coroutines.CoroutineName
-import kotlinx.coroutines.asCoroutineDispatcher
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.async
-import kotlinx.coroutines.delay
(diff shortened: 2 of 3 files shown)


kotlin-result-coroutines/src/jvmTest/kotlin/com/github/michaelbull/result/coroutines/binding/AsyncSuspendableBindingTest.kt
```diff
@@ -144,11 +145,17 @@ class AsyncSuspendableBindingTest {
...
+
         runBlocking {
             val result = binding<Unit, BindingError> {
-                launch(newThread("Thread A")) { provideX().bind() }
-                launch(newThread("Thread B")) { provideY().bind() }
-                launch(newThread("Thread C")) { provideZ().bind() }
+                launch(dispatcherA) { provideX().bind() }
+                dispatcherA.advanceTimeBy(20)
+                launch(dispatcherB) { provideY().bind() }
+                dispatcherB.advanceTimeBy(20)
```
Munzey, line 157: perhaps a bit contrived 😅  trying to represent the two threads running to 20 milliseconds at approx the same time, though in reality dispatcherA here only needs to be fast forwarded to 2

kotlin-result-coroutines/build.gradle.kts
```diff
@@ -31,6 +31,7 @@ kotlin {
             dependencies {
                 implementation(kotlin("test-junit"))
```
michaelbull, line 32: I think we can get rid of these now with https://blog.jetbrains.com/kotlin/2021/04/kotlin-1-5-0-rc-released/#test-library-changes

Might be worth making a separate PR though

Issue #64, opened 2021-10-05
Use of runCatching and coroutines breaks structured concurrency


grodin:
Since `runCatching` catches _all_ exceptions, it catches `CancellationException` which prevents any coroutines launched within the `runCatching` block from being cancelled when their enclosing `CoroutineScope` is cancelled.

This is not, strictly speaking, an issue with kotlin-result but it caught me by surprise so I thought it would probably be good to warn others of the potential gotcha.

I propose:

1. A warning in the kdoc comment, pointing out the problem or,
2. 1 + a new function in `kotlin-result-coroutines` called something like `runSuspendCatching()` which rethrows `CancellationException`.

I've already written a version of `runSuspendCatching()` and a test case that demonstrates the behaviour in my own project so I'm more than happy to submit a PR if you want.

The same issue has been raised for the stdlib `runCatching` at Kotlin/kotlinx.couroutines#1814 but it doesn't seem like the Kotlin devs thinks it's much of a problem. I disagree and thought maybe we could at least improve things here!

grodin:
See [this comment][1] on the above linked issue for an example of what goes wrong.

[1]: https://github.com/Kotlin/kotlinx.coroutines/issues/1814#issuecomment-588317835

michaelbull:
The existing signature of runCatching is likely to stay true to whatever the stdlib does, I don't think its wise to deviate from that in terms of user expectations. I do however like the idea proposed [here](https://github.com/Kotlin/kotlinx.coroutines/issues/1814#issuecomment-762701839) with the `except` function. Maybe we could make that work for more than just exceptions too. That also means that in future, the signature of `runSuspendCatching` would just be `runCatching { it }.except<CancellationException>()` or similar.

grodin:
That's quite funny: I had exactly the same thought and that's exactly how I implemented `runSuspendCatching` in my own project!

I'd suggest a slightly better name for `except()`, like `rethrow()` or something to make it a bit clearer what it does.

I'm not quite sure what you mean by making that extra function `rethrow` or whatever work for things other than exceptions. If the error isn't an exception how do we throw it? Did you have something else in mind.

Do you want separate PRs, firstly for `except`/`rethrow` and then another for `runSuspendCatching` to keep the git history nice or just one PR with unsquashed commits?

Munzey:
I also think a syntax like `rethrow()<>` for the specific result types where Err extends Exception reads better.

[...]

PR #66, opened 2021-10-11
Introduces throwIf and throwUnless

grodin:
Introduces the `rethrow` function as discussed in #64.


grodin:
@michaelbull I'll submit a PR for `runSuspendCatching` without this, and then if we find a way to make this PR work, we can rewrite `runSuspendCatching` in terms of `rethrow()` if it seems worth it.

grodin:
Actually, I've just remembered another potential solution to this I'd thought of. I'll just sketch it here: 

```kotlin
fun <V> Result<V, Throwable>.rethrowIf(
    predicate: (Throwable) -> Boolean
): Result<V, Throwable>
``` 

Then we could write

```kotlin
runSuspendCatching( block: suspend () ->V) = 
    runCatching(block)
        .rethrowIf { it is CancellableException }
```

PR #75, opened 2022-04-08
Add bind for nullables

pablisco:
⚠️ Need to add tests

So, we often find ourselves binding nullables with `toResultOr` + `bind`. We wanted to do an extension function, but until we get multiple receivers it's not gonna be possible. So I thought it was worth creating a default function to `ResultBinding`


michaelbull:
Could you give me an example of how this ends up being called?

Munzey:
I think it seems like a useful enough helper function to be included. If we go ahead with it, would you update the coroutines module as well please?

PR #86, opened 2023-05-09
toErrorIfNull to return Result with non-null value.

### LINKED ISSUE #84: ensureNotNull-alike binding scope extensions
Arrow has a handy extension for eagerly returning from effect with an error if a value is null - [`ensureNotNull`](https://arrow-kt.io/docs/apidocs/arrow-core/arrow.core.continuations/ensure-not-null.html). Would be very helpful to have a similar extension here, right now need to rely on fairly nested `flatMap`:

```kotlin
val value = maybeGetValue()
  .flatMap {
    when (it) {
      null -> Err(ValueMissingError)
      else -> Ok(it)
    }
  }
  .bind()
```

vs 

```kotlin
val value = maybeGetValue().bind()
ensureNotNull(value) { ValueMissingError }
```

kirillzh:
Resolves #84

### DIFF
--- a/kotlin-result/src/commonTest/kotlin/com/github/michaelbull/result/MapTest.kt
+++ b/kotlin-result/src/commonTest/kotlin/com/github/michaelbull/result/MapTest.kt
@@ -192,7 +192,7 @@ class MapTest {
 
         @Test
         fun returnsTransformedErrorIfNull() {
-            val result = Ok(null).toErrorIfNull { "a" }
+            val result: Result<Nothing, String> = Ok(null).toErrorIfNull { "a" }
 
             result as Err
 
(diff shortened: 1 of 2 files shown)


kotlin-result/src/commonTest/kotlin/com/github/michaelbull/result/MapTest.kt
```diff
@@ -192,7 +192,7 @@ class MapTest {
 
         @Test
         fun returnsTransformedErrorIfNull() {
-            val result = Ok(null).toErrorIfNull { "a" }
+            val result: Result<Nothing, String> = Ok(null).toErrorIfNull { "a" }
```
kirillzh, line 195: This would not have compiled before because the return type of `toErrorIfNull` was `Result<Nothing?, String>` which should not be possible.

michaelbull:
Merged in d7dbf35bcf305ea3dcfbe4f22e02b2fa5d6ee43d

michaelbull:
Released in v1.1.18

PR #98, opened 2024-03-09
Use List as return type for zipOrAccumulate

YuitoSato:
I unified the return type of some zipOrAccumulate functions from Result<V, Collection> to Result<V, List>. This was probably a mistake, but if it was intentional, please reject this.

Returning Result<V, Collection> is inconvenient because, even though the content is List, it requires casting or similar operations on the caller's side.


YuitoSato:
Thank you!

PR #112, opened 2024-10-28
Fix an issue with zipOrAccumulate with two procedures.

rhirai-line:
PR for #111 .

I modified zipOrAccumulate with two procedures.
Previously each procedure was executed twice, but now it is executed only once.

PR #113, opened 2024-11-02
Use Zulu instead of Temurin.

eichisanden:
Switch the Java distribution in CI from Temurin to Zulu.

The reason is that the current macos-latest runs on macOS 14 with arm64 architecture,
but Temurin doesn’t offer an arm64 build of Java 8.
https://github.com/actions/setup-java/issues/625

I became aware of this issue when I saw #112.
I hope this PR gets merged soon.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,7 +46,7 @@ jobs:
       - name: Setup JDK
         uses: actions/setup-java@9704b39bf258b59bc04b50fa2dd55e9ed76b47a8 # v4.1.0
         with:
-          distribution: temurin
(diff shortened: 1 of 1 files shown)


michaelbull:
Please could you make the appropriate change for the publish workflow as well? As the publish workflow doesn't run on pull requests I didn't notice that we've missed it and the master branch is now failing.

eichisanden:
>Please could you make the appropriate change for the publish workflow as well? As the publish workflow doesn't run on pull requests I didn't notice that we've missed it and the master branch is now failing.

I fixed this issue in #114.

Issue #104, opened 2024-06-03
Result 2.0: unsafe unwrapping of `Result.value` and `Result.error`


kirillzh:
I noticed that the new inline value class in [API 2.0](https://github.com/michaelbull/kotlin-result/releases/tag/2.0.0) lets us access result.value directly without checking if it’s actually safe, which wasn’t the case before. Here’s what I mean:

```kotlin
val result: Result<Int, SomeError> = getResult()
result.value // This compiles even if the result is an error, leading to potential runtime errors.
```

This could lead to runtime exceptions that were harder to make before when the API required type checking and more explicit unwrapping using `getOrThrow()` (which is a lot easier to catch or denylist through code analysis). 

It seems like API 2.0 introduces a downgrade in compile-time safety compared to the previous version. Could we consider reinstating some form of safety, perhaps through Kotlin contracts or another mechanism, to prevent potential runtime errors?

michaelbull:
> Could we consider reinstating some form of this safety, perhaps through Kotlin contracts or another mechanism, to prevent potential runtime errors?

Do you have a proposal of which compiler contract would achieve this? We effectively need user-defined guards (such that you can't call `.value` unless you've done an `isOk`/`isErr` check - I don't think such a thing exists in Kotlin?

raamcosta:
If `value` is always available, then maybe it should be nullable. This way we can check if it is null, and get smart cast.

It would also mean that people are more likely to use better approaches such as doing `.onSuccess { value -> }` given that in this case, "value" wouldn't be null.

But I might be missing something?

michaelbull:
It being nullable would force it to be boxed at all call sites. See the following example from Kotlin's docs:

```kotlin
interface I

@JvmInline
value class Foo(val i: Int) : I

fun asNullable(i: Foo?) {}

fun main() {
    val f = Foo(42)

    asNullable(f)  // boxed: used as Foo?, which is different from Foo
}
```

https://kotlinlang.org/docs/inline-classes.html#representation

raamcosta:
For us at work, in the type of apps I develop, I’d rather have safe access than the optimisations. So this means we’ll keep using version 1 then.

I guess it would be interesting to know how much users of the library fall into the same camp as us.

Also, I’m not sure if version 1 would need to have some support going forward? With updates to Kotlin versions and stuff like that? Probably not new features though.

kirillzh:
[...]

PR #123, opened 2025-04-27
Add UnsafeResultValueAccess annotation to Result class for safer value access

hoc081098:
Issue #104 

This pull request introduces a new `@UnsafeResultValueAccess` annotation to enforce safer access patterns for `Result.value` and `Result.error` in the `kotlin-result` library. Additionally, it updates the Kotlin conventions to opt into this annotation. Below are the key changes:

### Enhancements to `Result` class safety:

* Added the `@UnsafeResultValueAccess` annotation to mark access to `Result.value` and `Result.error` as potentially unsafe unless the `Result` state is explicitly checked using `Result.isOk` or `Result.isErr`. This encourages safer usage patterns. (`kotlin-result/src/commonMain/kotlin/com/github/michaelbull/result/Result.kt`, [[1]](diffhunk://#diff-84a8bfe8d281dcc7d79384c9770c83586c852246035e81d93eb2f0cd7d14cc02R54-R59) [[2]](diffhunk://#diff-84a8bfe8d281dcc7d79384c9770c83586c852246035e81d93eb2f0cd7d14cc02R107-R122)

### Kotlin conventions update:

* Updated the Kotlin conventions in `kotlin-conventions.gradle.kts` to opt into the `@UnsafeResultValueAccess` annotation, ensuring that projects using this configuration can leverage the new safety mechanism. (`buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts`, [buildSrc/src/main/kotlin/kotlin-conventions.gradle.ktsR62](diffhunk://#diff-512f24ca000c4a4211f731f7eae50d295161039d8c8018e5a9ee61d3a8182f33R62))

### DIFF
--- a/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts
@@ -59,6 +59,7 @@ kotlin {
         all {
             languageSettings.apply {
                 optIn("kotlin.contracts.ExperimentalContracts")
+                optIn("com.github.michaelbull.result.UnsafeResultValueAccess")
             }
         }
 
--- a/kotlin-result/src/commonMain/kotlin/com/github/michaelbull/result/Result.kt
+++ b/kotlin-result/src/commonMain/kotlin/com/github/michaelbull/result/Result.kt
@@ -51,10 +51,12 @@ public value class Result<out V, out E> internal constructor(
     private val inlineValue: Any?,
 ) {
 
+    @UnsafeResultValueAccess
     @Suppress("UNCHECKED_CAST")
     public val value: V
         get() = inlineValue as V
 
+    @UnsafeResultValueAccess
     @Suppress("UNCHECKED_CAST")
     public val error: E
         get() = (inlineValue as Failure<E>).error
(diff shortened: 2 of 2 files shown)


michaelbull:
This is a great idea! Can you show me what it looks like in IntelliJ when you try to use the `.value` and `.error` fields without explicitly adding the annotation? Does it cause a compilation error or is it a warning?

michaelbull:
That's a really nice solution. Is there a way that we can point users to a compiler flag that turns this annotation off, such that we don't break all their existing code if they aren't ready to adopt it?

hoc081098:
> That's a really nice solution. Is there a way that we can point users to a compiler flag that turns this annotation off, such that we don't break all their existing code if they aren't ready to adopt it?

We absolutely want to avoid breaking existing code for users who aren't ready to adopt this yet. According to [Kotlin's opt-in documentation](https://kotlinlang.org/docs/opt-in-requirements.html), we can point users to explicitly opt in by [annotating their usage with @OptIn(UnsafeResultValueAccess::class)](https://kotlinlang.org/docs/opt-in-requirements.html#opt-in-locally), or [configure it globally via Gradle](https://kotlinlang.org/docs/opt-in-requirements.html#opt-in-a-module).

We'll include this in the changelog and provide guidance on both approaches so users can adopt the change at their own pace.

michaelbull:
That's great - I really appreciate the detailed explanation and how much thought you've put into this. I have to investigate the build failures on GitHub Actions before I can approve the workflow to be built.

Issue #124, opened 2025-04-28
'expect'/'actual' classes (including interfaces, objects, annotations, enums, and 'actual' typealiases) are in Beta.


hoangchungk53qx1:
Emit the compilation warning on expect/actual classes. The warning must mention that expect/actual classes are in Beta

![Image](https://github.com/user-attachments/assets/b024ef9d-9c20-4d9c-b9c3-c541135edcf0)

michaelbull:
Can this be done with an `optIn`, the same way we do for the [`ExperimentalContracts`](https://github.com/michaelbull/kotlin-result/blob/6603671a2c851b6e46c4a884f5626e89d6249951/buildSrc/src/main/kotlin/kotlin-conventions.gradle.kts#L63) - or do we have to do it with `compilerArgs`?

hoangchungk53qx1:
I have just created PR to this, pls, check my PR

michaelbull:
Addressed in dcb85a60ad5f42c9da0f34258aceb568159111c8

PR #125, opened 2025-04-28
Fix Expect Actual warning

hoangchungk53qx1:
Fix issues : https://github.com/michaelbull/kotlin-result/issues/124

Emit the compilation warning on expect/actual classes. The warning must mention that expect/actual classes are in Beta


michaelbull:
Merged in dcb85a60ad5f42c9da0f34258aceb568159111c8, thanks!

PR #135, opened 2025-12-29
feat: unused return value checker

rileymichael:
closes #134 as discussed by enabling the unused return value checker in `full` mode for the entire project and explicitly marking methods where the return value is safe to ignore with `@IgnorableReturnValue` (`onSuccess` and `onFailure`).

i've also enabled the return value checker in `check` mode in the example application to validate. here is an example warning (note that there are no warnings in the existing usage):

5A8wzEA_d

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -6,7 +6,7 @@ pluginManagement {
 }
 
 plugins {
-    id("org.gradle.toolchains.foojay-resolver-convention") version "0.8.0"
+    id("org.gradle.toolchains.foojay-resolver-convention") version "1.0.0"
 }
 
 dependencyResolutionManagement {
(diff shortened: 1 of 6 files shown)


settings.gradle.kts
```diff
@@ -6,7 +6,7 @@ pluginManagement {
 }
 
 plugins {
-    id("org.gradle.toolchains.foojay-resolver-convention") version "0.8.0"
+    id("org.gradle.toolchains.foojay-resolver-convention") version "1.0.0"
```
rileymichael: gradle 9 removed a [deprecated jvm vendor](https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#removal_of_deprecated_jvmvendorspec_ibm_semeru) which broke compatibility with foojay toolchain resolver versions [before 1.0.0](https://github.com/gradle/foojay-toolchains/blob/main/CHANGELOG.md#changed-1)

